PR #21, opened 2017-08-17
 Enhancements to support new use cases. Documentation fixes.

rajive:
Hi Javi:

I have made some enhancements to support the ability to more easily plugin a new Qos profiles without duplication of XML files, fixed some documentation errors, and also removed some hard-coded dependencies.

1. New feature: -qosLibrary <library name>

- Allows users to plugin custom qos profiles
- Custom profiles can extend the perftest builtin profile (without having to duplicate into a new copy)
- Custom profile can be an application specific profile, eg custom transports etc.
- Can easily compare application specific profile performance vs. optimized perftest qos profile
- Created example file to show how this is to be used

2. Added a new "TransportQos" profile to make it easy to plugin custom transport configurations.

3. Fixed some hard-coded assumptions:

- Now if a transport is not specified, the default as specified in the XML QoS profile is used. Previously, it was assumed to be UDPv4. Future: Should probably add a -enableUdp command line parameter to override the default in XML.

- Now if -bestEffort is not specified, the default setting for RELIABILITY comes from the XML profile.
Future: Should probably add a -reliable command line parameter to override the default in XML.

- Removed some internal hard-coded magic strings with variables that can be overridden via command line arguments

4. Fixed several documentation errors, inconsistencies, and clarified certain items.

More details can be found in the git log.

---

To Do:

- Renerate the qos_string.h from the updated perftest_qos_profiles.xml
 
Thanks
Rajive

### DIFF
--- a/my_perftest_qos_profiles.xml
+++ b/my_perftest_qos_profiles.xml
@@ -1,0 +1,101 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
+
+<!--
+(c) 2005-2017  Copyright, Real-Time Innovations, Inc. All rights reserved.
+Subject to Eclipse Public License v1.0; see LICENSE.md for details.
+-->
+
+<!--
+This file contains the QoS configurations used by the RTI PerfTest, a
+performance test for measuring the latency of the middleware at different
+throughput levels.
+
+The format of this file is described in the RTI Connext Core Libraries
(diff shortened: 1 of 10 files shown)


javmorcas:
The changes proposed by Rajive in this pull request have been addressed [here|https://github.com/rticommunity/rtiperftest/pull/25]

Issue #59, opened 2018-04-05
Unreliable behavior when using Best Effort + Scan in certain architectures


javmorcas:
We found that using the combination of Best Effort + Scan + LatencyTest in architectures where the CPU is limited could cause the test to hang, take a lot to run and report an incorrect number of sample lost.

The reason seems to be here:

```
                    message.size = LENGTH_CHANGED_SIZE;
                    // must set latency_ping so that a subscriber sends us
                    // back the LENGTH_CHANGED_SIZE message
                    message.latency_ping = num_pings % _NumSubscribers;
                    while (announcement_reader_listener->announced_subscribers > 0) {
                        writer->Send(message, true);
                        writer->Flush();
                        writer->wait_for_acknowledgments(
                                timeout_wait_for_ack_sec,
                                timeout_wait_for_ack_nsec);
                    }
```

When using best effort, the `wait_for_acknowledgments()` call will return immediately, and that loop will be continuously exercised until `announcement_reader_listener->announced_subscribers` is 0. However that will only happen when all the subscribers answer with a message in the "Announcement" topic, the fact that we are in a busy loop and flooding the network will affect negatively to the reception of that message. This is specially noticeable when pub and sub apps are in the same machine and that machine does not have much CPU.

Associated with this issue, the samples with the LENGTH_CHANGED_SIZE that are sent and not received will be accounted as sample lost, which should not be the case.

PR #60, opened 2018-04-06
Unreliable behavior when using Best Effort + Scan in certain architectures (#59)

javmorcas:
Pull request for #59.

### DIFF
--- a/srcCpp/RTIDDSImpl.cxx
+++ b/srcCpp/RTIDDSImpl.cxx
@@ -784,6 +784,7 @@ class RTIPublisher : public IMessagingWriter
     DDS_InstanceHandle_t *_instance_handles;
     RTIOsapiSemaphore *_pongSemaphore;
     long _instancesToBeWritten;
+    bool _isReliable;
 
  public:
     RTIPublisher(DDSDataWriter *writer, unsigned long num_instances, RTIOsapiSemaphore * pongSemaphore, int instancesToBeWritten)
@@ -792,7 +793,7 @@ class RTIPublisher : public IMessagingWriter
         data.bin_data.maximum(0);
         _num_instances = num_instances;
         _instance_counter = 0;
-        _instance_handles = 
+        _instance_handles =
                 (DDS_InstanceHandle_t *) malloc(sizeof(DDS_InstanceHandle_t)*(_num_instances + 1)); // One extra for MAX_CFT_VALUE
         _pongSemaphore = pongSemaphore;
         _instancesToBeWritten = instancesToBeWritten;
@@ -808,6 +809,10 @@ class RTIPublisher : public IMessagingWriter
             data.key[c] = (unsigned char)(MAX_CFT_VALUE >> c * 8);
         }
         _instance_handles[_num_instances] = _writer->register_instance(data);
+
+        DDS_DataWriterQos qos;
+        _writer->get_qos(qos);
+        _isReliable = (qos.reliability.kind == DDS_RELIABLE_RELIABILITY_QOS);
     }
 
     ~RTIPublisher() {
@@ -931,9 +936,13 @@ class RTIPublisher : public IMessagingWriter
         return (unsigned int)status.pulled_sample_count;
     };
 
-    void wait_for_acknowledgments(long sec, unsigned long nsec) {
-        DDS_Duration_t timeout = {sec, nsec};
-        _writer->wait_for_acknowledgments(timeout);
+    void waitForAck(long sec, unsigned long nsec) {
+        if (_isReliable) {
+            DDS_Duration_t timeout = {sec, nsec};
+            _writer->wait_for_acknowledgments(timeout);
+        } else {
+            MilliSleep(nsec / 1000000);
+        }
     }
 };
 
@@ -997,6 +1006,10 @@ public:
             fprintf(stderr, "set_octet_array(key) failed: %d.\n", retcode);
         }
         _instance_handles[_num_instances] = _writer->register_instance(data);
+
+        DDS_DataWriterQos qos;
+        _writer->get_qos(qos);
+        _isReliable = (qos.reliability.kind == DDS_RELIABLE_RELIABILITY_QOS);
     }
 
     ~RTIDynamicDataPublisher() {
@@ -1173,9 +1186,13 @@ public:
         return (unsigned int)status.pulled_sample_count;
     };
 
-    void wait_for_acknowledgments(long sec, unsigned long nsec) {
-        DDS_Duration_t timeout = {sec, nsec};
-        _writer->wait_for_acknowledgments(timeout);
+    void waitForAck(long sec, unsigned long nsec) {
+        if (_isReliable) {
+            DDS_Duration_t timeout = {sec, nsec};
+            _writer->wait_for_acknowledgments(timeout);
+        } else {
+            MilliSleep(nsec / 1000000);
+        }
     }
 };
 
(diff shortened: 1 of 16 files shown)


srcCpp/perftest_publisher.cxx
```diff
@@ -1836,10 +1849,14 @@ int perftest_cpp::Publisher()
...
-                            timeout_wait_for_ack_nsec);
-                    announcement_reader_listener->announced_subscribers = _NumSubscribers;
+                    if (_IsReliable) {
+                        writer->wait_for_acknowledgments(
+                                timeout_wait_for_ack_sec,
+                                timeout_wait_for_ack_nsec);
+                    } else {
+                        MilliSleep(timeout_wait_for_ack_nsec / 1000000);
+                    }
+                    announcement_reader_listener->announced_subscriber_replies = _NumSubscribers;
```
fercsrti77: There is something I don't understand. Here we assign announced_subscriber_replies to _NumSubscribers. However, in AnnouncementListener::ProcessMessage we increase that value. I do not understand the logic of increasing versus setting to _NumSubscribers

srcCpp/perftest_publisher.cxx
```diff
@@ -1850,12 +1867,27 @@ int perftest_cpp::Publisher()
...
+                     * lost, and potentially CPU starbation for other processes.
+                     * We can prevent this by adding a small sleep() if the test is best
+                     * effort.
+                     */
+
+                    while (announcement_reader_listener->announced_subscriber_replies > 0) {
                         writer->Send(message, true);
                         writer->Flush();
-                        writer->wait_for_acknowledgments(
+                        if (_IsReliable) {
```
fercsrti77: Instead of implementing this distinction here, I would move the logic inside wait_for_acknowledgements. perftest_publisher is supposed to be middleware agnostic and here you are making assumptions based on reliability. In addition, it will save the If in multiple places

javmorcas: I agree that would be a saving, however it would require to add to the RTIPublisher class a variable to check if the publisher is reliable or not (since, currently we don't have access to that information, do you want to proceed that way? @fercsrti77

srcCpp/perftest_publisher.cxx
javmorcas: @fercsrti77, @softwarejimenez and I realized about the flaws of the current implementation and decided to modify the announcement reader logic using some code suggested by @softwarejimenez. The Pull request has been updated with the new changes. We can go through the changes tomorrow.

srcCpp/perftest_publisher.cxx
```diff
@@ -136,8 +136,9 @@ int perftest_cpp::Run(int argc, char *argv[])
 }
 
 // Set the default values into the array _scanDataLenSizes vector
-const void set_default_scan_values(
-        std::vector<unsigned long> & _scanDataLenSizes){
+void set_default_scan_values(
```
javmorcas: No need to use const

srcCpp/perftest_publisher.cxx
```diff
@@ -1143,10 +1144,22 @@ int perftest_cpp::Subscriber()
...
+     * different things.
+     *
+     * We will use 3 sizes: INITIALIZE_SIZE, LENGTH_CHANGED_SIZE and FINISHED_SIZE,
+     * msg.data will be used as the payload of such messages, so we choose the
+     * greatest size:
+     */
+    TestMessage announcement_msg;
+    announcement_msg.entity_id = _SubID;
+    announcement_msg.size = perftest_cpp::INITIALIZE_SIZE;
+    announcement_msg.data = new char[LENGTH_CHANGED_SIZE];
```
javmorcas: LENGTH_CHANGED_SIZE is the greatest size of all 3. Alternatively we could do max(a,max(b,c))

srcCpp/perftest_publisher.cxx
```diff
@@ -1702,7 +1734,8 @@ int perftest_cpp::Publisher()
     // indicating that it has discovered every Publisher
     fprintf(stderr,"Waiting for subscribers announcement ...\n");
     fflush(stderr);
-    while (_NumSubscribers > announcement_reader_listener->announced_subscribers) {
+    while (_NumSubscribers
+            > (int)announcement_reader_listener->subscriber_list.size()) {
```
javmorcas: This is the first usecase: We are waiting until we receive the information of all the readers (we know we must receive _NumSubscribers number of them.

srcCpp/perftest_publisher.cxx
```diff
@@ -1836,10 +1872,13 @@ int perftest_cpp::Publisher()
...
-                            timeout_wait_for_ack_sec,
-                            timeout_wait_for_ack_nsec);
-                    announcement_reader_listener->announced_subscribers = _NumSubscribers;
+                    if (_IsReliable) {
+                        writer->wait_for_acknowledgments(
+                                timeout_wait_for_ack_sec,
+                                timeout_wait_for_ack_nsec);
+                    } else {
+                        MilliSleep(timeout_wait_for_ack_nsec / 1000000);
+                    }
```
javmorcas: We still need to decide if we want to keep this approach or move the if(_isReliable) to the writer wait_for_acknowledgments call. This would require the writer to know about its reliability (which makes sense, but it is not propagated to the writer in the current implementation.

@fercsrti77

srcCpp/perftest_publisher.cxx
```diff
@@ -1850,12 +1889,28 @@ int perftest_cpp::Publisher()
...
+                     * a wait for acknowledgements. However, if the Throughput topic is
+                     * Best Effort, wait_for_acknowledgments() will return inmediately.
+                     * This would cause that the Send() would be exercised too many times,
+                     * in some cases causing the network to be flooded, a lot of packets being
+                     * lost, and potentially CPU starbation for other processes.
+                     * We can prevent this by adding a small sleep() if the test is best
+                     * effort.
+                     */
+                    announcement_reader_listener->subscriber_list.clear();
+                    while ((int)announcement_reader_listener->subscriber_list.size()
```
javmorcas: We do:  message.size = LENGTH_CHANGED_SIZE;
We clear the list of subscribers. and then we send the messages, since these are answered via the announcement writer of the Perftest subscribers, that will trigger that we will reload the list, and once that list is equal to the number of expected subscribers, that means all the Subscribers know about the change in datalen

srcCpp/perftest_publisher.cxx
javmorcas: @fercsrti77, I implemented this option but then I also implemented your suggestion of using the waitForAck function. I believe the second option makes more sense.

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2519,11 +2519,11 @@ IMessagingReader *RTIDDSImpl<T>::CreateReader(
     }
 
     dr_qos.resource_limits.initial_instances = _InstanceCount + 1;
-    if (_InstanceMaxCountReader != DDS_LENGTH_UNLIMITED) {
+    if (_InstanceMaxCountReader != (unsigned long)DDS_LENGTH_UNLIMITED) {
```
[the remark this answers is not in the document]
javmorcas: Fixed

srcCpp/perftest_publisher.cxx
```diff
@@ -1921,14 +1976,17 @@ int perftest_cpp::Publisher()
 
     message.size = FINISHED_SIZE;
     unsigned long i = 0;
-    announcement_reader_listener->announced_subscribers = _NumSubscribers;
-    while (announcement_reader_listener->announced_subscribers > 0
+    while (announcement_reader_listener->subscriber_list.size() > 0
             && i < initializeSampleCount) {
```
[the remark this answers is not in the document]
javmorcas: @fercsrti77 This makes sense to me as well, I discussed this with antonio and we modified a bit these variables (see new changes)

PR #71, opened 2018-05-03
Maximum Latency is not correctly obtained when the clock has low resolution

javmorcas:
Issue #58

### DIFF
--- a/srcCpp/perftest_publisher.cxx
+++ b/srcCpp/perftest_publisher.cxx
@@ -1356,7 +1356,7 @@ class LatencyListener : public IMessagingCB
         latency_sum = 0;
         latency_sum_square = 0;
         count = 0;
-        latency_min = 0;
+        latency_min = ULONG_MAX;
         latency_max = 0;
         last_data_length = 0;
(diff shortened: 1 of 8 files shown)


srcCpp/perftest_publisher.cxx
```diff
@@ -1356,7 +1356,7 @@ class LatencyListener : public IMessagingCB
         latency_sum = 0;
         latency_sum_square = 0;
         count = 0;
-        latency_min = 0;
+        latency_min = ULONG_MAX;
```
javmorcas: [Jose]: Initialize as a sentinel instead of directly to ULONG_MAX.

srcCpp/perftest_publisher.cxx
```diff
@@ -1521,19 +1521,14 @@ class LatencyListener : public IMessagingCB
             }
         }
 
-        if (latency_min == 0)
-        {
+        if (latency_min == ULONG_MAX) {
```
javmorcas: [Jose]: Initialize as a sentinel instead of directly to ULONG_MAX.

PR #98, opened 2018-06-04
Fix percentage so it fits in the same line as the rest of info

javmorcas:


javmorcas:
This needs to be cherrypicked into release/2.4

Issue #109, opened 2018-06-26
Remove a unnecessary TODO in java


alfonso-soto:
**Description**
In rtiperftest/srcJava/com/rti/perftest/harness/LatencyListener.java:205

```
// TODO: This std dev calculation isn't correct!
double latency_std = sqrt(
        abs(_latencySumSquare / (double)_count -
                (latency_ave * latency_ave)));
```
**The formula is correct and the TODO is not needed.**

**TODO**:
- [x] Support for Java
- [x] Full [Code Review]()
- [ ] Final Merge into master

PR #110, opened 2018-06-26
Unnecessary TODO removed

alfonso-soto:
This pull request is coming from the issue #109 . When the pull request is merged into "master" the issue will be [resolved, closed, fixed] #109 

After reviewing the code we concluded that the formula is correct, therefore this comment is unnecessary and we are removing it.


javmorcas:
After reviewing the code we concluded that the formula is correct, therefore this comment is unnecessary and we are removing it.

Issue #111, opened 2018-07-02
Logic error on RTIDynamicDataSubscriber constructor.


alfonso-soto:
**Description**

In rtiperftest/srcCpp/RTIDDSImpl.cxx:1684.
Possible call to NULL pointer.

`
    RTIDynamicDataSubscriber(DDSDataReader *reader): _message()
    {
        _reader = DDSDynamicDataReader::narrow(reader); //fail -> _reader = NULL
        if (_reader == NULL) {
            fprintf(stderr,"DDSDynamicDataReader::narrow(reader) error.\n");
           // throw runtime_error exception
        }
        _data_idx = 0;
        _no_data = false;

        // null listener means using receive thread
        if (_reader->get_listener() == NULL) { //Call to NULL pointer FAIL
            ...
            ...
        }
    }
`
If  DDSDynamicDataReader::narrow(reader) fail, and return NULL, a call to NULL pointer will be done for _reader->get_listener() where _reader will be NULL.

**Expected behavior**

A exception should be throw if  `_reader == NULL` after call `narrow(reader)`

**TODO**:
- [x] Support for C++ Traditional
- [ ] First CR [Code Review]()
- [ ] Documentation
- [ ] Full [Code Review]()
- [ ] Final Merge into master

PR #112, opened 2018-07-02
Removing Logic error on RTIDynamicDataSubscriber constructor.

alfonso-soto:
This pull request is coming from the issue #111 . When the pull request is merged into "master" the issue will be [resolved, closed, fixed] #111 

**Initial Problem**

Running Clang Static Analyzer, we found a logic error as follow:

In rtiperftest/srcCpp/RTIDDSImpl.cxx:1684.
Possible call to NULL pointer.

```
RTIDynamicDataSubscriber(DDSDataReader *reader): _message()
{
    _reader = DDSDynamicDataReader::narrow(reader); //fail -> _reader = NULL
    if (_reader == NULL) {
        fprintf(stderr,"DDSDynamicDataReader::narrow(reader) error.\n");
        // throw runtime_error exception
    }
    _data_idx = 0;
    _no_data = false;

    // null listener means using receive thread
    if (_reader->get_listener() == NULL) { //Call to NULL pointer FAIL
        ...
        ...
    }
}

```
If `DDSDynamicDataReader::narrow(reader)` fails, and `return NULL`, a call to `NULL` pointer will be done for `_reader->get_listener()` where `_reader` will be `NULL`, ending with a segmentation fault.

**Solution**
 Throwing an exception each time the function `narrow` is called by a constructor, calling the shutdown function before.

**New problem**
When a function called by `Publisher()` or `Subscriber()` fails, a `return` is done without destroying any local variables, some of them created with `alloc` or `new`.

```
    // create latency pong _writer
    _writer = _MessagingImpl->CreateWriter(LATENCY_TOPIC_NAME);

    if (_writer == NULL) {
        fprintf(stderr, "Problem creating latency _writer.\n");
        return -1;
    }
```
**Solution**
All these variables have been moved as `private` inside of `perftest_cpp.h` with their corresponding destructor sentence.

This came up with a dependencies problem, where these new variables inside of perftest_cpp.h were objects of classes defined on perftest_publisher.cxx 
We decided to follow the standard way and create new files with these classes.

### DIFF
--- a/srcCpp/RTIDDSImpl.cxx
+++ b/srcCpp/RTIDDSImpl.cxx
@@ -1167,9 +1167,11 @@ public:
     }
 
     void Shutdown() {
-        if (_writer->get_listener() != NULL) {
-            delete(_writer->get_listener());
-            _writer->set_listener(NULL);
+        if (_writer != NULL) {
+            if (_writer->get_listener() != NULL) {
+                delete(_writer->get_listener());
+                _writer->set_listener(NULL);
+            }
         }
 
         free(_instance_handles);
@@ -1606,12 +1608,14 @@ class RTISubscriber : public IMessagingReader
 
     void Shutdown()
     {
-        if (_reader->get_listener() != NULL) {
-            delete(_reader->get_listener());
-            _reader->set_listener(NULL);
+        if (_reader != NULL) {
+            if (_reader->get_listener() != NULL) {
+                delete(_reader->get_listener());
+                _reader->set_listener(NULL);
+            }
+            // loan may be outstanding during shutdown
+            _reader->return_loan(_data_seq, _info_seq);
         }
-        // loan may be outstanding during shutdown
-        _reader->return_loan(_data_seq, _info_seq);
     }
 
     TestMessage *ReceiveMessage()
--- a/srcCpp/RTIDDSImpl.cxx
+++ b/srcCpp/RTIDDSImpl.cxx
@@ -111,7 +111,7 @@ void RTIDDSImpl<T>::Shutdown()
 
         DDSDomainParticipantListener* participant_listener = _participant->get_listener();
         if (participant_listener != NULL) {
-            delete(participant_listener);
+            delete participant_listener;
         }
 
         _participant->delete_contained_entities();
@@ -957,7 +957,7 @@ class RTIPublisher : public IMessagingWriter
     void Shutdown() {
         if (_writer != NULL) {
             if (_writer->get_listener() != NULL) {
-                delete(_writer->get_listener());
+                delete _writer->get_listener();
                 _writer->set_listener(NULL);
             }
         }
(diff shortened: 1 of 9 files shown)


srcCpp/perftestReadersListeners.h
```diff
@@ -0,0 +1,141 @@
...
+public:
+    ThroughputListener(
+            IMessagingWriter *writer,
+            IMessagingReader *reader = NULL,
+            bool UseCft = false,
+            int numPublishers = 1);
+
+    ~ThroughputListener();
+
+    IMessagingReader *get_reader();
```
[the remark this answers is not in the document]
alfonso-soto: Not really possible for `message = listener->get_reader()->ReceiveMessage();`

message is a non const variable.

srcCpp/RTIDDSImpl.cxx
```diff
@@ -917,6 +908,17 @@ class RTIPublisher : public IMessagingWriter
     RTIPublisher(DDSDataWriter *writer, unsigned long num_instances, RTIOsapiSemaphore * pongSemaphore, int instancesToBeWritten)
     {
         _writer = T::DataWriter::narrow(writer);
+        if (_writer == NULL) {
+            fprintf(stderr, "DataWriter::narrow(writer) error.\n");
+            /*
+             * Nothing was created at this point, shutdown() is called for
+             * maintain consistency.
+             */
+            Shutdown();
```
alexcamposruiz: _instance_handles hasn't been initialized when Shutdown calls free on it. If _instance_handles is garbage, free() can crash. 

Initialize _instance_handles to NULL or don't call Shutdown().

srcCpp/perftestReadersListeners.h
```diff
@@ -0,0 +1,146 @@
+/*
```
alexcamposruiz: I think this file is not part of this code review. I will skip it.

srcCpp/perftest_publisher.cxx
```diff
@@ -176,31 +176,59 @@ void set_default_scan_values(
...
     }
 
     if (_MessagingArgv != NULL) {
-        delete []_MessagingArgv;
+        delete[] _MessagingArgv;
     }
 
-    if(_MessagingImpl != NULL){
+    if (_throughputReaderListener != NULL) {
+        delete (_throughputReaderListener);
```
alexcamposruiz: Don't use parenthesis around delete: ```delete _throughputReaderLlistener```

srcCpp/perftestReadersListeners.h
```diff
@@ -0,0 +1,146 @@
...
+};
+
+
+/*
+ * perftestReaderListener.cxx have some dependencies from perftest_cpp.h.
+ * Also perftest_cpp.h needs the tree classes defined here.
+ * As solution, the include of perftest_cpp.h has been moved to the
+ * end of this file.
+ */
+#include "perftest_cpp.h"
```
alfonso-soto: Actually, I would like to know what do you think about the include that I need to add at the end of the file. @alexcamposruiz 

Other solution is place it on the .cxx file.

javmorcas:
Given all the changes in code, I am closing this issue. We will work on this in a new Pull Request once we come back to it.

PR #119, opened 2018-08-02
Add spin and spin per microseconds support for micro

alfonso-soto:
Spin and SpinPerMicroSecond implementation for micro.

### DIFF
--- a/srcCpp/Infrastructure_micro.h
+++ b/srcCpp/Infrastructure_micro.h
@@ -159,21 +159,6 @@ class NDDSUtility {
     #endif
     };
 
-    /* Get the time cost to call getTimeUsec a number of times equal to iteration */
-    static unsigned long long get_time_overhead(unsigned int iterations) {
-
-        PerftestClock clock = PerftestClock::getInstance();
-        unsigned long long init = 0, fin = 0, overhead = 0;
-
-        init = clock.getTimeUsec();
-        for (int i = 0; i < iterations; i++) {
-            clock.getTimeUsec();
-        }
-        fin = clock.getTimeUsec();
-
-        return fin - init;
-    }
-
     /*e \dref_Utility_spin */
     static void spin(DDS_UnsignedLongLong spinCount){
         NDDS_Utility_spin(spinCount);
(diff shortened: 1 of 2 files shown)


srcCpp/Infrastructure_micro.h
```diff
@@ -159,16 +159,59 @@ class NDDSUtility {
...
+        PerftestClock clock = PerftestClock::getInstance();
+        unsigned long long init = 0, fin = 0, overhead = 0;
+
+        init = clock.getTimeUsec();
+        for (int i = 0; i < iterations; i++) {
+            clock.getTimeUsec();
+        }
+        fin = clock.getTimeUsec();
+
+        return fin - init;
```
javmorcas: I believe overhead is not used, right? In that case it can be deleted. then you can just declare 

```
unsigned long long init = clock.getTimeUsec();
```

and then:

```
return clock.getTimeUsec() - init;
```

srcCpp/Infrastructure_micro.h
```diff
@@ -159,16 +159,59 @@ class NDDSUtility {
...
+        /* Same default values used by DDS */
+        unsigned int spinCount = 20000;
+
+        unsigned long long init = 0, fin = 0, usec = 0;
+        unsigned long long iterations = 0;
+        unsigned long long overhead = 0;
+
+        PerftestClock clock = PerftestClock::getInstance();
+
+        while((int)(usec - overhead) < (int)precision){
```
javmorcas: Add a comment explaining why you use this (basically what you explained to me)

srcCpp/Infrastructure_micro.h
```diff
@@ -159,16 +159,59 @@ class NDDSUtility {
...
+            usec += fin - init;
+
+            iterations++;
+            overhead = get_time_overhead(iterations);
+        }
+
+        /* Measure the clock.getTimeUsec() overhead */
+        usec -= overhead;
+
+        return (float)(iterations * spinCount) / usec;
```
javmorcas: Add a ward for usec, in certain OS it can be "0", due to unreliable clocks with low resolution

javmorcas:
I added some suggestions. @alfonso-soto, could you review them?

PR #128, opened 2018-08-21
Fix issues with Dynamic Data C99 compatibility

javmorcas:


javmorcas:
@softwarejimenez, could you review this? I believe this is the change you mentioned in the issue.

PR #139, opened 2019-01-10
Allow the use of the legacy Dynamic Data implementation

javmorcas:


build.bat
```diff
@@ -98,6 +98,8 @@ if NOT "%1"=="" (
 				SET RELEASE_DEBUG=debug
 		) ELSE if "%1"=="--dynamic" (
 				SET STATIC_DYNAMIC=dynamic
+		) ELSE if "%1"=="--legacy-DynamicData" (
+				SET LEGACY_DD_IMPL=1
```
softwarejimenez: define before LEGACY_DD_IMPL:

```
set LEGACY_DD_IMPL=0
```

build.sh
```diff
@@ -635,6 +641,9 @@ while [ "$1" != "" ]; do
         --secure)
             USE_SECURE_LIBS=1
             ;;
+        --legacy-DynamicData)
+            LEGACY_DD_IMPL=1
```
softwarejimenez: define the variable first
```
LEGACY_DD_IMPL=0
```

build.bat
javmorcas: Thanks! Good point!

build.bat
```diff
@@ -256,6 +258,11 @@ if !BUILD_CPP! == 1 (
 		set "additional_defines_custom_type= -D RTI_CUSTOM_TYPE=%custom_type%"
 	)
 
+	if !LEGACY_DD_IMPL! == 1 (
+        echo [INFO]: Allow the use of both legacy and new Dynamic Data Impl.
```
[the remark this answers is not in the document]
javmorcas: I think the problem is that we are using tabs and here I added spaces. Fixed.

build.sh
javmorcas: Fixed

javmorcas:
@softwarejimenez, I added your CR feedback

PR #144, opened 2019-03-09
Added support for Security with Micro

asorbini:
In order to compile, the generated CMakeLists.txt must be modified to link rti_me_seccore, ssl, and crypto (or sslz, and cryptoz, if linking statically), beside adding OPENSSLHOME/lib to the linker path

This is of course not the "final solution", but I'm not familar with what additional changes are required to modify the rtiddsgen-generated files (I tried passing additional aguments in build.sh but that didn't seem to be picked up).


javmorcas:
Added changes in #179.

PR #149, opened 2019-03-13
Support Large Data and Shared Memory in Micro branch

javmorcas:


build.bat
```diff
@@ -427,6 +431,12 @@ if !BUILD_MICRO! == 1 (
 
 	call::solution_compilation_flag_calculation
 
+	if !BUILD_MICRO_24x_COMPATIBILITY! == 1 (
+		set "ADDITIONAL_DEFINES=RTI_MICRO_24x_COMPATIBILITY !ADDITIONAL_DEFINES!"
+	) else (
+		set "rtiddsgen_extra_options=!rtiddsgen_extra_options! -additionalRtiLibraries nddsmetp"
```
javmorcas: Is this to support zero copy or flat data? Is this required for shared memory? @franciscoRTI

srcIdl/perftest.idl
```diff
@@ -5,6 +5,7 @@
 
 const long MAX_SYNCHRONOUS_SIZE = 63000;
 const long MAX_BOUNDED_SEQ_SIZE = 63000;
+const long TMP_MICRO_MAX_UNBOUNDED_SEQ_SIZE = 1048576; // 1024*1024
```
javmorcas: Did we find the solution for this at the end? Didn't you fixed the issue in the templates for rtiddsgen? @franciscoRTI

javmorcas: I mean to pass the "unbounded" thing and so.

build.bat
franciscoRTI: yes, for some reason this is the name that Pro uses; in our codegen templates we translate it to the actual library names that are in Micro. This will be updated to the actual library names here in the branch with micro security support for codegen.

srcIdl/perftest.idl
franciscoRTI: Yes, I think we can remove that and leave the sequences without a size. Latest micro codegen correctly should support -stringSize and -sequenceSize, but I have not tested that.

PR #159, opened 2019-04-25
Raw Transport merge

javmorcas:
Until we merge the parameter manager branch to master, we will use this as a CR for the changes for Raw Transport.


javmorcas:
@alfonso-soto will update this branch to add support for 6.0.0 (the serialization function changed between these versions).

PR #168, opened 2019-05-07
Feature/152 build script compiler linker params

### LINKED ISSUE #152: Improve build script for selecting compiler/linker
**Description**
Currently the compiler and linker used by the build script can be set through the `COMPILER` and `LINKER` environment variables. However, this is not documented. This tripped me and a customer up during cross-compilation (e.g. for the armv8 platform we had to use `aarch64-linux-gnu-g++` when compiling on Ubuntu). I recommend we either improve the documentation or add command line parameters to the build script e.g. `-compilerName` and `-linkerName` so as to self-document this capability through the `-help` string.

NOTE: The 5th bullet point [in the prerequisites section of the compilation docs](https://github.com/rticommunity/rtiperftest/blob/master/srcDoc/compilation.rst#prerequisites) recommends having the compiler/linker parent directory in the PATH variable, but it does not state that the executable name can be changed through environment variables.

**TODO**:
- [ ] Support for C++ Traditional
- [ ] First CR [Code Review]()
- [ ] Support for C++ Modern
- [ ] Support for Java
- [ ] Support for CS
- [ ] Automatic install testing
- [ ] Documentation
- [ ] Full [Code Review]()
- [ ] Final Merge into master

KyleBenson:
This pull request is coming from the issue #152. When the pull request is merged into "master" the issue will be resolved #152


javmorcas:
@KyleBenson I had no comments with regards to the Code Changes. This looks great to me, also good documentation.

I forgot to mention that you will need to add the issue to the Release Notes page. Just add it to the bottom of the "What's new" section of the Release Notes `.rst` file. Use the #152 for the number of the issue (you will see more examples there).

Great work!

PR #175, opened 2019-05-24
Feature/166/fix shared parameters

### LINKED ISSUE #166: In certain VxWorks when using publisher and subscriber in the same machine the -noPrint parameter is shared
**Description**

In certain VxWorks when using publisher and subscriber in the same machine the -noPrint parameter is shared

**How to Reproduce**

**Expected behavior**

**TODO**:
- [x] Support for C++ Traditional
- [x] First CR [Code Review]()
- [x] Support for C++ Modern
- [x] Automatic install testing
- [x] Documentation
- [ ] Full [Code Review]()
- [ ] Final Merge into master

salvacorts:
This pull request is coming from the issue #166 . When the pull request is merged into "master" the issue will be closed and fixed #166


salvacorts:
Since VxWorks loads the same shared object, some addresses, like static class members, are shared across instances of the program within the same machine.

I have removed those static members by making them instance members. I think it would be worth having a look at other potentially problematic static variables; so far I have took care of `showCpu` (*cpu* param), `subID` (*sidMultiSubTest* param) and `printIntervals` (noPrint param), but there are other static variables like `perftest_cpp::INITIALIZE_SIZE` for example that I don't know if they are problematic in this context.

This is an screenshot of two perftest instances running in the same VxWorks machine, one with noPrint parameter (the subscriber, right), and one without (the publisher, left).
Screenshot from 2019-05-24 11-19-41

srcCpp/Infrastructure_common.h
```diff
@@ -14,42 +14,31 @@
   #include "Infrastructure_pro.h"
 #endif
 
+#ifdef RTI_VXWORKS
```
javmorcas: I think we might not need this if we make nddsUtility:sleep() work properly in VxWorks (this belongs to the alarm issue though).

srcCpp/Infrastructure_pro.cxx
```diff
@@ -5,6 +5,19 @@
 
 #include "Infrastructure_pro.h"
 
+// Since std::to_string is not defined in c++98
```
javmorcas: If the comment uses more than 1 line:

/*
 * line 1
 * line 2
 */

srcCpp/Infrastructure_pro.cxx
```diff
@@ -5,6 +5,19 @@
...
+// we will define it here for VxWorks
+#ifdef RTI_VXWORKS
```
javmorcas: I think this might be use as well for Micro, i would move it to infrastructure_common, what do you think?

srcCpp/Infrastructure_pro.h
```diff
@@ -16,6 +16,10 @@
   #include "PerftestSecurity.h"
 #endif
 
+#ifdef RTI_VXWORKS
```
javmorcas: Same as previous comment

srcCpp/Infrastructure_common.h
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
salvacorts: Sure, but for what I have seen so far,  `infrastructure_common.h` includes `infrastructure_[micro | pro].h` How will Micro access this then?

salvacorts: I have tried to move it to `Infrastructure_common.cxx` and it has failed compiling

salvacorts:
You should be able to merge it now Javi

PR #179, opened 2019-05-28
Merge Micro Support code

javmorcas:
This pull request is coming from the issue #XX. When the pull request is merged into "master" the issue will be [resolved, closed, fixed] #XX


javmorcas:
After reviewing this code with Francis, from the Micro team and checking that performance is not degraded with respect to 2.4. I will proceed to do the merge.

PR #191, opened 2019-06-18
Feature/flatdata support

salvacorts:
Support for _FlatData_ language binding and _Zero Copy_ over shared memory in _rtiperftest_

- [x] Support for Traditional C++
- [x] Support for Modern C++
- [ ] Testing suite
- [x] Benchmark results


.vscode/launch.json
```diff
@@ -30,22 +29,48 @@
             "MIMode": "lldb",
         },
         {
-            "name": "Pro Debug g++",
+            "name": "Pro Debug g++ (Publisher)",
             "type": "cppdbg",
             "request": "launch",
-            "program": "${workspaceFolder}/bin/${architecture}/debug/perftest_cpp",
-            "args": ["-pub", "-domain", "0", "-latencyCount", "1", "-dataLen", "1000", "-latencyTest", "-multicast", "-executionTime", "2", "-noPrint"],
+            "program": "${workspaceFolder}/bin/x64Linux4gcc7.3.0/debug/perftest_cpp03",
```
javmorcas: @salvacorts, make sure you change back this to the ${architecture} variable, and maybe create a copy for "Pro Debug g++ (Publisher Cpp03)"

.vscode/launch.json
```diff
@@ -30,22 +29,48 @@
...
+                    "ignoreFailures": true
+                }
+            ]
         },
         {
             "name": "Micro Debug g++",
             "type": "cppdbg",
             "request": "launch",
-            "program": "${workspaceFolder}/bin/${architecture}/debug/perftest_cpp_micro",
+            "program": "${workspaceFolder}/bin/x64Linux4gcc7.3.0/debug/perftest_cpp_micro",
```
javmorcas: same as before

build.bat
```diff
@@ -284,12 +292,13 @@ if !BUILD_CPP! == 1 (
...
 	call::copy_src_cpp_common
 	call::solution_compilation_flag_calculation
+	call::get_flatdata_available
 
 	REM # Generate files for the custom type files
 	set "additional_defines_custom_type="
 	set "additional_header_files_custom_type="
 	set "additional_source_files_custom_type="
-
+
```
javmorcas: Remove spaces

build.bat
```diff
@@ -334,6 +343,7 @@ if !BUILD_CPP! == 1 (
 	)
 
 	set "ADDITIONAL_DEFINES=RTI_LANGUAGE_CPP_TRADITIONAL"
+	set additional_rti_libs=
```
javmorcas: I think there is no need to add this line

srcCpp03/perftest_publisher.cxx
```diff
@@ -117,15 +117,55 @@ int perftest_cpp::Run(int argc, char *argv[]) {
...
         if (_PM.get<bool>("keyed")) {
-            _MessagingImpl = new RTIDDSImpl<TestDataKeyed_t>();
+            if (_PM.get<bool>("flatdata")) {
+                #ifdef RTI_FLATDATA_AVAILABLE
+                 if (_PM.get<bool>("zerocopy")) {
+                    _MessagingImpl = new RTIDDSImpl_FlatData<TestDataKeyed_ZeroCopy_w_FlatData_t>();
+                 } else {
+                    _MessagingImpl = new RTIDDSImpl_FlatData<TestDataKeyed_FlatData_t>();
+                 }
+                #endif //RTI_FLATDATA_AVAILABLE
```
alexcamposruiz: #else, give an error

srcCpp03/RTIDDSImpl.h
```diff
@@ -38,7 +38,7 @@ class RTIDDSImpl : public IMessaging
 {
   public:
 
-    RTIDDSImpl();
+    RTIDDSImpl(bool isFlatData=false);
```
alexcamposruiz: 1) Why do you pass isFlatData as a constructor argument, when all the other options are part of that `_PM` type.
2) If you still pass this argument, the constructor must be `explicit` to disable implicit conversion from `bool` to `RTIDDSImpl`.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -80,10 +80,11 @@ int DynamicDataMembersId::at(std::string key)
 }
 
 template <typename T>
-RTIDDSImpl<T>::RTIDDSImpl():
+RTIDDSImpl<T>::RTIDDSImpl(bool isFlatData):
         _InstanceMaxCountReader(dds::core::LENGTH_UNLIMITED), //(-1)
```
alexcamposruiz: why is `//(-1)` a relevant comment?

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -205,6 +206,15 @@ bool RTIDDSImpl<T>::validate_input()
                 _PM->set<long>("batchSize", 0); // Disable Batching
             }
         }
+
+        if (_isFlatData) {
+            if (_PM->is_set("batchSize") && _PM->get<long>("batchSize") > 0) {
+                fprintf(stderr, "Batching cannot be used with FlatData.\n");
```
alexcamposruiz: We should be using `std::cerr`, but I see we use fprintf in the rest of the file.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -205,6 +206,15 @@ bool RTIDDSImpl<T>::validate_input()
...
+                return false;
+            } else {
+                _PM->set<long>("batchSize", -3);
```
alexcamposruiz: what is -3? If it has a special meaning, use a constant variable.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -555,7 +580,7 @@ class RTIPublisher: public RTIPublisherBase<T> {
             this->data.key()[c] = (unsigned char) (key >> c * 8);
         }
         if (!isCftWildCardKey) {
-            this->_writer.write(this->data, this->_instance_handles[key]);
+            this->_writer.write(this->data, this->_instance_handles.back());
```
alexcamposruiz: the way to select the instance handle calls for a comment. This code is not self-descriptive.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -564,6 +589,110 @@ class RTIPublisher: public RTIPublisherBase<T> {
...
+
+  public:
+      RTIFlatDataPublisher(
+              dds::pub::DataWriter<T> writer,
+              int num_instances,
+              rti::core::Semaphore& pongSemaphore,
+              bool useSemaphore,
+              int instancesToBeWritten,
+              ParameterManager *PM
+              )
```
alexcamposruiz: coding conventions: the placement of `)` and `:` is incorrect:

```
RTIFlatDataPublisher(
    ...)
    : RTIPublisher...
```

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -564,6 +589,110 @@ class RTIPublisher: public RTIPublisherBase<T> {
...
+      }
+
+  public:
+      RTIFlatDataPublisher(
+              dds::pub::DataWriter<T> writer,
+              int num_instances,
+              rti::core::Semaphore& pongSemaphore,
+              bool useSemaphore,
+              int instancesToBeWritten,
+              ParameterManager *PM
```
alexcamposruiz: why do we use `PM` everywhere and not `parameterManager`?

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -564,6 +589,110 @@ class RTIPublisher: public RTIPublisherBase<T> {
...
+
+          add_key(builder, key);
+
+          // Build the data to be sent
+          T *sample = builder.finish_sample();
+
+          // Send data through the writer
+          if (!isCftWildcardKey) {
+              // .back() because the CFT Handler is always the last one to be pushed back
+              this->_writer.write(*sample, this->_instance_handles.back());
```
alexcamposruiz: Minor enhancement: instead of writing a comment to indicate this invariant, make it explicit in code. For example, define a private member function `get_cft_instance_handle() { return _instance_handles.back(); }`

Don't worry about performance; the compiler will inline it.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -564,6 +589,110 @@ class RTIPublisher: public RTIPublisherBase<T> {
...
+          add_key(builder, MAX_CFT_VALUE);
+          T *sample = builder.finish_sample();
+
+          this->_instance_handles.push_back(
+              this->_writer.register_instance(*sample));
+
+          this->_writer.extensions().discard_loan(*sample);
+      }
+
+      inline bool send(TestMessage &message, bool isCftWildcardKey) {
```
alexcamposruiz: Don't use `inline`. `inline` is used for non-member functions whose body is defined in a header. In a member function `inline` is unnecessary and compilers ignore it.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -1719,61 +1818,178 @@ IMessagingReader *RTIDDSImpl<T>::CreateReader(
...
+    if (!_PM->get<bool>("dynamicData")) {
+        dds::topic::Topic<T> topic(_participant, topic_name);
+        dds::sub::DataReader<T> reader(dds::core::null);
+        ReceiverListener<T> *reader_listener = NULL;
 
-    // only force reliability on throughput/latency topics
-    if (topic_name != ANNOUNCEMENT_TOPIC_NAME.c_str()) {
-        if (!_PM->get<bool>("bestEffort")) {
-            qos_reliability = dds::core::policy::Reliability::Reliable();
+        if (topic_name == THROUGHPUT_TOPIC_NAME.c_str() && _PM->is_set("cft")) {
```
alexcamposruiz: This is a pointer comparison, not a string comparison. Is that what you want?

Even if you do, the full string comparison (`topic_name == THROUGHPUT_TOPIC_NAME`) is more robust and maintainable.

build.bat
salvacorts: True, I was doing it because in line 357 I was doing: `set additional_rti_libs=nddssecurity !additional_rti_libs!` so I wanted to initialize `additional_rti_libs` somehow. I can safely remove it.

srcCpp03/perftest_publisher.cxx
salvacorts: Why do you want to give an error if RTI_FLATDATA_AVAILABLE is not defined?

I have wrapped those lines in a `#ifdef RTI_FLATDATA_AVAILABLE` in case rtiperftest is compiled against a ConnextDDS version previous to 6.0.0. In that case since flatdata is not available, all the related flatdata code wont be compiled.

Does the later make sense?

srcCpp03/RTIDDSImpl.h
salvacorts: I have removed that argument and initialized it to _false_ in the constructor for `RTIDDSImpl` and to _true_ in `RTIDDSImpl_FlatData`.

```suggestion
template <typename T>
RTIDDSImpl<T>::RTIDDSImpl():
        _InstanceMaxCountReader(dds::core::LENGTH_UNLIMITED), //(-1)
        _InstanceHashBuckets(dds::core::LENGTH_UNLIMITED), //(-1)
        _isLargeData(false),
        _isFlatData(false),
        ...

...

template <typename T>
RTIDDSImpl_FlatData<T>::RTIDDSImpl_FlatData() {
    this->_isFlatData = true;
}
```

srcCpp03/RTIDDSImpl.cxx
salvacorts: I do not know, it was there before I committed my changes. I guess it is to clarify what is the actual value of `dds::core::LENGTH_UNLIMITED` (which is -1).

Nevertheless, I think `dds::core::LENGTH_UNLIMITED` is self-explanatory and therefore `(-1)` should be removed.

I will wait until @javmorcas gives feedback about it

srcCpp03/RTIDDSImpl.cxx
salvacorts: I have changed all `printf(stderr)` to `std::cerr`.

srcCpp03/RTIDDSImpl.cxx
salvacorts: Yes, it indicates that batchSize argument is disabled because FlatData is active. I was copying this behavior from line 192
```C++
if (_isLargeData) {
            if (_PM->is_set("batchSize") && _PM->get<long>("batchSize") != 0) {
                std::cerr << "Batching cannot be used with Large Data." 
                          << std::endl;
                return false;
            } else {
                _PM->set<long>("batchSize", -2);
            }
} else ...
```

These codes are used in `perftest_publisher.cpp` to display a message in the "header" of perftest explaining the reason why the batchsize argument is disabled. 

```C++
     if (_PM.get<long>("batchSize") == -1) {
           if (_PM.get<bool>("latencyTest")) {
               stringStream << "\t\t  BatchSize disabled for a Latency Test\n";
           } else {
               stringStream << "\t\t  BatchSize is smaller than 2 times\n"
                            << "\t\t  the sample size.\n";
           }
       } else if (_PM.get<long>("batchSize") == -2) {
           stringStream << "\t\t  BatchSize cannot be used with\n"
                        << "\t\t  Large Data.\n";
        } else if (_PM.get<long>("batchSize") == -3) {
            stringStream << "\t\t  BatchSize cannot be used with\n"
                         << "\t\t  FlatData.\n";
        }
```

I will speak with @javmorcas about refactoring this.

srcCpp03/RTIDDSImpl.cxx
salvacorts: Corrected

srcCpp03/RTIDDSImpl.cxx
salvacorts: I was following the naming used in the rest of the file. I can change it if @javmorcas also agrees. Even though in my opinion PM is self-explanatory enough, I agree that `parameterManager` is a more descriptive name.

srcCpp03/RTIDDSImpl.cxx
salvacorts: Ok, good idea!

srcCpp03/RTIDDSImpl.cxx
salvacorts: Good to know, thank you!

srcCpp03/RTIDDSImpl.cxx
salvacorts: I moved that code from `CreateReader` function to `setup_DR_QoS` but this comparison was there previous to my commits, maybe @javmorcas knows more about that. 

As far as I understand we should rather not do a pointer comparison there since in case we hardcode the `topic_name` argument when calling `setup_DR_QoS` function, this comparison will not work as desired. I agree that string comparison is a better approach.

 I will change this and other similar cases on my next commit.

salvacorts:
I have committed changes from this CR.

perftest_qos_profiles.xml
```diff
@@ -250,8 +250,8 @@ _KeepDurationUsec variable name corresponds to the command-line option -keepDura
 
       </participant_qos>
 
-      <!-- === Publisher Configuration: ================================= 
-
```
javmorcas: If we are finally not changing this file, could you revert all the space changes?

srcCpp03/RTIDDSImpl.cxx
javmorcas: I agree with you 2.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -235,7 +261,7 @@ bool RTIDDSImpl<T>::validate_input()
 
     // Manage transport parameter
     if (!_transport.validate_input()) {
-         std::cerr << "[Error] Failure validation the transport options."
+         std::cerr << "[Error] Failure validation the transport options."
```
javmorcas: Remove extra space

build.sh
```diff
@@ -544,12 +581,16 @@ function build_cpp()
     ##############################################################################
     # Generate files for srcCpp
 
+    # rtiddsgen ignores any specified rti addional library if using ZeroCopy
+    # Therefore, we need to generate a makefile that contains
+    # nddsmetp and nddssecurity libraries
```
javmorcas: I think this comment is the same as the one below.

perftest_qos_profiles.xml
```diff
@@ -250,8 +250,8 @@ _KeepDurationUsec variable name corresponds to the command-line option -keepDura
 
       </participant_qos>
 
-      <!-- === Publisher Configuration: ================================= 
-        
+      <!-- === Publisher Configuration: =================================
```
javmorcas: Revert this file since we are not touching it.

srcIdl/perftest.idl
```diff
@@ -39,9 +39,9 @@ struct TestData_t {
 
 }; //@Extensibility FINAL_EXTENSIBILITY
 
-
 struct TestDataKeyed_t {
-    octet key[KEY_SIZE]; //@key
+    @key
```
javmorcas: Check if this compiles for 5.3.0, else, keep the old notation.

srcCpp03/perftest_publisher.cxx
javmorcas: I think what this means is that if you use -flatdata and you don't have flat data available, then, the test will finish silently. It should show a message saying that flat data is not available for that perftest version.

Also, we need to check how this works when using micro.

build.bat
```diff
@@ -351,12 +360,19 @@ if !BUILD_CPP! == 1 (
 				echo [ERROR]: In order to link statically using the security plugin you need to also provide the OpenSSL home path by using the --openssl-home option.
 				exit /b 1
 			)
-			set rtiddsgen_extra_options=-additionalRtiLibraries nddssecurity -additionalLibraries "libeay32z ssleay32z"
+			set additional_rti_libs=nddssecurity !additional_rti_libs!
```
javmorcas: Make sure no spaces are there if !addition_rti_libs! is empty, or the resulting solution will complain.

build.bat
```diff
@@ -367,6 +383,16 @@ if !BUILD_CPP! == 1 (
 	set "additional_header_files=!additional_header_files_custom_type!!additional_header_files!RTIRawTransportImpl.h Parameter.h ParameterManager.h ThreadPriorities.h RTIDDSLoggerDevice.h MessagingIF.h RTIDDSImpl.h perftest_cpp.h qos_string.h CpuMonitor.h PerftestTransport.h Infrastructure_common.h Infrastructure_pro.h"
 	set "additional_source_files=!additional_source_files_custom_type!!additional_source_files!RTIRawTransportImpl.cxx Parameter.cxx ParameterManager.cxx ThreadPriorities.cxx RTIDDSLoggerDevice.cxx RTIDDSImpl.cxx CpuMonitor.cxx PerftestTransport.cxx Infrastructure_common.cxx Infrastructure_pro.cxx"
 
+	@REM # TODO: REMOVE THIS
```
javmorcas: We cannot have a comment like this in the code. I believe we decided that for supporting codegen for 6.0.0 we need this, right? This is the same we are doing in the .sh

build.bat
```diff
@@ -375,13 +401,46 @@ if !BUILD_CPP! == 1 (
...
 	!rtiddsgen_extra_options! !additional_defines_custom_type!^
 	-d "%classic_cpp_folder%" "%idl_location%\perftest.idl"
 	if not !ERRORLEVEL! == 0 (
 		echo [ERROR]: Failure generating code for %classic_cpp_lang_string%.
 		call::clean_src_cpp_common
 		exit /b 1
 	)
+
+	if !FLATDATA_AVAILABLE! == 1 (
+		@REM # TODO: REMOVE THIS
```
javmorcas: Add a comment instead explaining why we do this, check with Aida if we have a bug ID in Jira for this, else, make sure we create a Jira issue and add here that Jira issue.

build.bat
javmorcas: I belive we only need to do it twice right? not 3 times.

build.bat
```diff
@@ -416,20 +478,37 @@ if !BUILD_CPP03! == 1 (
 				echo [ERROR]: In order to link statically using the security plugin you need to also provide the OpenSSL home path by using the --openssl-home option.
 				exit /b 1
 			)
-			set rtiddsgen_extra_options=-additionalRtiLibraries nddssecurity -additionalLibraries "libeay32z ssleay32z"
+			set additional_rti_libs=nddssecurity !additional_rti_libs!
```
javmorcas: Apply same comments as in C++ traditional

build.bat
```diff
@@ -732,73 +854,75 @@ GOTO:EOF
...
+	echo.                           	  not present, the path to PERL should be
+	echo.                           	  available from your \$PATH variable.
+	echo.    --java-home path       	  Path to the Java JDK home folder. If this
+	echo.                           	  parameter is not present, javac, jar and java
+	echo.                           	  executables should be available from your
+	echo.                           	  $PATH variable.
+	echo.                           	  (No effect when building for Micro)
+	echo.    --debug                	  Compile against the RTI Connext Debug
+	echo.                           	  libraries. Default is against release ones.
+	echo.    --dynamic                    Compile against the RTI Connext Dynamic
```
javmorcas: Check indentation.

build.bat
```diff
@@ -732,73 +854,75 @@ GOTO:EOF
...
+	echo.                           	  Note: For Micro provide this path with /
+	echo.                           	  instead of \, required by cmake.
+	echo.    --clean                	  If this option is present, the build.sh script
+	echo.                           	  will clean all the generated code and binaries
+	echo.                           	  from previous executions.
+	echo.    --customType type      	  Use the Custom type feature with your type.
+	echo.                           	  See details and examples of use in the
+	echo.                           	  documentation.
+	echo.    --flatdata-max-size size     Specify the maximum bounded size in bytes      
+    echo.                                 for sequences when using FlatData language
```
javmorcas: Indentation is done with tabs, but separation for the echos with spaces.

build.sh
```diff
@@ -758,11 +823,19 @@ function build_micro_cpp()
...
 {
     copy_src_cpp_common
+    check_flatdata_available
     additional_defines_calculation "CPPModern"
+
     ##############################################################################
     # Generate files for srcCpp03
-
-    rtiddsgen_command="\"${rtiddsgen_executable}\" -language ${modern_cpp_lang_string} -unboundedSupport -replace -create typefiles -create makefiles -platform ${platform} -additionalHeaderFiles \"ThreadPriorities.h Parameter.h ParameterManager.h MessagingIF.h RTIDDSImpl.h perftest_cpp.h qos_string.h CpuMonitor.h PerftestTransport.h\" -additionalSourceFiles \"ThreadPriorities.cxx Parameter.cxx ParameterManager.cxx RTIDDSImpl.cxx CpuMonitor.cxx PerftestTransport.cxx\" -additionalDefines \"${additional_defines}\" ${rtiddsgen_extra_options} -d \"${modern_cpp_folder}\" \"${idl_location}/perftest.idl\""
+    rtiddsgen_command="\"${rtiddsgen_executable}\" -language ${modern_cpp_lang_string} \
```
javmorcas: Check the code in master, I believe we reworked this, you will have to do some changes here.

javmorcas: Do a merge with the latest changes from master.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -84,7 +97,8 @@ RTIDDSImpl<T>::RTIDDSImpl():
         _InstanceMaxCountReader(dds::core::LENGTH_UNLIMITED), //(-1)
```
javmorcas: Apply feedback from Alex.

srcCpp03/RTIDDSImpl.cxx
javmorcas: Create an issue for after 3.0.0 to use constant names instead of the plain numbers.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -413,6 +458,14 @@ class RTIPublisherBase: public IMessagingWriter {
     bool _isReliable;
     ParameterManager *_PM;
 
+     /* Returns the instance handler for the content filtered topic.
```
javmorcas: Multi-line comments should look like:

```
/*
 * Line 1
 * Line 2
 */
```

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -548,7 +600,7 @@ class RTIPublisher: public RTIPublisherBase<T> {
                 }
             }
         } else {
-            key = MAX_CFT_VALUE;
+            key = this->_instance_handles.size()-1;
```
javmorcas: Can we double check that this is working properly, not sure why we used MAX_CFT_VALUE, and its value.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -557,13 +609,111 @@ class RTIPublisher: public RTIPublisherBase<T> {
...
         } else {
-            this->_writer.write(this->data,
-                    this->_instance_handles[this->_num_instances]);
+            this->_writer.write(this->data, this->getCftInstanceHandle());
         }
         return true;
     }
 };
 
+#ifdef RTI_FLATDATA_AVAILABLE
```
javmorcas: Add a comment explaining a bit about this class, its use and why we use it differently.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -1238,7 +1534,7 @@ dds::core::QosProvider RTIDDSImpl<T>::getQosProviderForProfile(
 
     if (!_PM->get<bool>("noXmlQos")) {
         qosProvider = dds::core::QosProvider(
-                _PM->get<std::string>("qosFile").c_str(),
+                _PM->get<std::string>("qosFile"),
```
javmorcas: Was this working before? Was this a bug? If so, I want to report it and add in the documentation (Release Notes) that it is fixed.

srcCpp03/RTIDDSImpl.cxx
```diff
@@ -1427,178 +1723,20 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const std::string &topic_name)
...
-    dw_qos << qos_reliability;
-    dw_qos << qos_resource_limits;
-    dw_qos << qos_dw_resource_limits;
-    dw_qos << qos_durability;
-    dw_qos << dwPublishMode;
-    dw_qos << dwBatch;
-    dw_dataWriterProtocol.rtps_reliable_writer(dw_reliableWriterProtocol);
-    dw_qos << dw_dataWriterProtocol;
-    dw_qos << Property(properties.begin(), properties.end(), true);
+    dds::pub::qos::DataWriterQos dw_qos = setup_DW_QoS(qos_profile, topic_name);
```
javmorcas: I like this.

build.sh
javmorcas: Forget about my last comment, but still, instead of listing the files, you can use the variable where the files are, like in cpp:

```
    additional_header_files=" \
        ThreadPriorities.h \
        Parameter.h \
        ParameterManager.h \
        MessagingIF.h \
        RTIDDSImpl.h \
        perftest_cpp.h \
        CpuMonitor.h \
        PerftestTransport.h \
        PerftestSecurity.h \
        Infrastructure_common.h \
        Infrastructure_micro.h"

    additional_source_files=" \
        ThreadPriorities.cxx \
        Parameter.cxx \
        ParameterManager.cxx \
        RTIDDSImpl.cxx \
        CpuMonitor.cxx \
        PerftestTransport.cxx \
        PerftestSecurity.cxx \
        Infrastructure_common.cxx \
        Infrastructure_micro.cxx"
```

build.sh
salvacorts: Done

perftest_qos_profiles.xml
salvacorts: Done

build.bat
salvacorts: I forgot to remove that. I was printing the command to be executed to debug it. But now that we know it works I will remove it.

build.bat
salvacorts: Done

build.sh
salvacorts: Done

srcCpp03/RTIDDSImpl.cxx
salvacorts: Done

srcCpp03/RTIDDSImpl.cxx
salvacorts: Done

srcCpp03/RTIDDSImpl.cxx
salvacorts: I am pretty sure this is working as expected. The problem is that I do not know how was that even working before since we were accessing an unallocated memory segment.

Before this change we were doing:
```C++
// In RTIPublisher constructor
for (unsigned long i = 0; i < this->_num_instances; ++i) { // this->_num_instances = 1 by default
    // Register the key of Instance
    this->_instance_handles.push_back(this->_writer.register_instance(this->data));
}

// Register the key of MAX_CFT_VALUE
this->_instance_handles.push_back(this->_writer.register_instance(this->data));

// ...

// In send function
// ...
key = MAX_CFT_VALUE; // MAX_CFT_VALUE = 65535 = [255,255,0,0,]b
//...
if (!isCftWildCardKey) {
    this->_writer.write(this->data, this->_instance_handles[key]);
}
```

So we would be doing `this->_instance_handles[key]` that is `this->_instance_handles[MAX_CFT_VALUE]` which is not accessible since _instance_handles has only two elements, not 65535.

This was causing me some errors during the execution of perftest. With these changes, it works.

srcCpp03/RTIDDSImpl.cxx
salvacorts: Done for every class and method of new FlatData classes

srcCpp03/RTIDDSImpl.cxx
salvacorts: Yes, it was working because `dds::core::QosProvider` gets two `std::string` as parameters.
You can assign a `const char *` to a string. It will convert it to an actual string.

build.bat
salvacorts: So far I have not had any complaint from the tool about having `" "` or `"nddssecurity "`. Did you?

srcCpp03/perftest_publisher.cxx
salvacorts: Rather thank giving an error; do not allow using `-flatData` as an option if it is not available:
```C++
    if (_PM.get<bool>("keyed")) {
          #ifdef RTI_FLATDATA_AVAILABLE
            if (_PM.get<bool>("flatdata")) {
                 if (_PM.get<bool>("zerocopy")) {
                    _MessagingImpl = new RTIDDSImpl_FlatData<TestDataKeyed_ZeroCopy_w_FlatData_t>(true);
                 } else {
                    _MessagingImpl = new RTIDDSImpl_FlatData<TestDataKeyed_FlatData_t>();
                 }
            } else {
                _MessagingImpl = new RTIDDSImpl<TestDataKeyed_t>();
            }
          #else 
            _MessagingImpl = new RTIDDSImpl<TestDataKeyed_t>();
          #endif //RTI_FLATDATA_AVAILABLE
        }
```

srcIdl/perftest.idl
salvacorts: Yes it does.

srcCpp03/RTIDDSImpl.cxx
salvacorts: Done

srcIdl/perftest.idl
```diff
@@ -80,8 +82,172 @@ struct TestDataKeyedLarge_t {
...
+
+#ifdef RTI_FLATDATA_AVAILABLE
+/*
+ * FLATDATA
+ */
+@mutable
+@language_binding(FLAT_DATA)
+struct TestData_FlatData_t {
+    octet key[KEY_SIZE];
+
```
harishkumaruk: unnecessary new line in all FlatData types in this file

srcIdl/perftest.idl
```diff
@@ -80,8 +82,172 @@ struct TestDataKeyedLarge_t {
...
+};
+
+@mutable
+@language_binding(FLAT_DATA)
+struct TestDataKeyed_FlatData_t {
+    @key
+    octet key[KEY_SIZE];
+
+    long entity_id;
```
harishkumaruk: mentioned above

srcIdl/perftest.idl
salvacorts: Done

srcIdl/perftest.idl
salvacorts: Done

srcCppCommon/ParameterManager.cxx
```diff
@@ -338,6 +338,38 @@ void ParameterManager::initialize()
             | Middleware::RTIDDSMICRO);
     create("threadPriorities", threadPriorities);
 
+  #ifdef RTI_FLATDATA_AVAILABLE
```
alfonso-soto: I don't see a problem allowing the creation of these parameters out of `RTI_FLATDATA_AVAILABLE`

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
-    // Number of messages that can be buffered in the receive queue.
-    int received_message_count_max = 1024 * 1024 * 2
-            / (int) _PM->get<unsigned long long>("dataLen");
-    if (received_message_count_max < 1) {
-        received_message_count_max = 1;
-    }
+    DDS_Property_t *parentProp = DDSPropertyQosPolicyHelper::lookup_property(qos.property,
+                "dds.transport.shmem.builtin.parent.message_size_max");
+    int parent_msg_size_max = atoi(parentProp->value);
+    std::cout << "message_size_max: " << parent_msg_size_max << std::endl;
```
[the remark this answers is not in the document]
salvacorts: These `couts` are for debugging purposes. They are not meant to be present at release.

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+     * So that, we need to find out the maximum allocable space to avoid
+     * runtime errors.
+     *
+     * Perform an incremental search of allocable space in range (minSize, maxSize).
+     */
+    RTIOsapiSharedMemorySegmentHandle handle;
+    RTI_UINT64 pid = RTIOsapiProcess_getId();
+    RTIBool success = RTI_FALSE;
+    int retcode;
+    int key = rand();
```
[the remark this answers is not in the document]
salvacorts: As far as I have seen; No, we are not. Should I do it at the beginning of the execution of perftest (`perftest_cpp::Run`)?

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+    int minSize = parent_msg_size_max;
+    int step = 1048576 + parent_msg_size_max; // 1MB + parent_msg_size_max
+    int maxBufferSize = 60817408 + step; // 58MB
```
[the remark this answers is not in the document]
salvacorts: Minimum size is the packet of the builtin transport for SHMEM. We need it as the minimum size for a SHMEM segment since otherwise, DDS would not even be able to send a sample.

The maximum size is there in order to have a limit on the buffer size for SHMEM so Perftest does not require unlimited resources.

Since we are doing a decremental search, we use a step of 1MB to search down a pseudo-optimal buffer size in the range [minSize, maxBufferSize].

I will add this explanation as a comment on the code.

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+
+    do {
+        maxBufferSize -= step;
```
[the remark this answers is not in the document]
salvacorts: Ops :man_facepalming:. Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+
+        // Reset handles to known state
+        RTIOsapiMemory_zero(&handle,
+                sizeof(struct RTIOsapiSharedMemorySegmentHandle));
+
+        success = RTIOsapiSharedMemorySegment_create(
```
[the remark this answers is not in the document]
salvacorts: `dds.transport.shmem.builtin.parent.message_size_max` is the maximum size of a message in bytes that can be sent or received by the transport plugin.

We need to know the maximum allocable space on SHMEM to allocate the receiving buffer (`dds.transport.shmem.builtin.receive_buffer_size`).

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2284,35 +2553,48 @@ double RTIDDSImpl<T>::obtain_dds_deserialize_time_cost(
...
-    DDS_DataWriterQos dw_qos;
-    std::string qos_profile;
-
-    DDSTopic *topic = _participant->create_topic(
-                       topic_name,
-                       _typename,
-                       DDS_TOPIC_QOS_DEFAULT,
-                       NULL,
-                       DDS_STATUS_MASK_NONE);
+unsigned long int RTIDDSImpl<T>::getShmemSHMMAX() {
```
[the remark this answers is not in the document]
salvacorts: Good idea :+1:. Done

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2495,6 +2779,369 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
...
+                            "Consider increasing SHMMAX parameter on your system settings or select a different transport.\n"
+                            "See https://community.rti.com/kb/what-are-possible-solutions-common-shared-memory-issues\n"
+                            "If you still run into this issue, consider cleaning your Shared Memory segments.\n");
+                }
+            }
+          #endif
+
+            // The writer_loaned_sample_allocation is initial_simples + 1
+            initial_samples = std::max(
+                    1ul, (max_allocable_space - RTI_FLATDATA_MAX_SIZE) / RTI_FLATDATA_MAX_SIZE);
```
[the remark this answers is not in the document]
salvacorts: No, but now we do it by checking at compile time since we also need to check this to compile IDLs.

srcCppCommon/ParameterManager.cxx
salvacorts: It would not make sense to allow the user to use the parameter `-flatdata` if their DDS version does not provide FlatData support.

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2495,6 +2779,369 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
...
+                max_allocable_space = getShmemSHMMAX();
+                fprintf(stdout, "SHMMAX: %lu\n", max_allocable_space);
+                //max_allocable_space = MAX_DARWIN_SHMEM_SIZE;
+
+                /**
+                 * Leave enought room for an sceneario of two participants:
+                 *   - One Publisher with one DW (throughput topic)
+                 *   - One Subscriber with two DW (Latency topic and Announcement)
+                 */
+                max_allocable_space /= 3;
```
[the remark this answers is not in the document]
salvacorts: Because it is the only SHMEM resources constrained OS we are supporting. This is done in order to allow perftest running on OSX with low SHMEM resources even though we lose performance. (We still display a message about how to increase these resource limits)

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+                &handle, &retcode, key, maxBufferSize, pid);
+
+        RTIOsapiSharedMemorySegment_delete(&handle);
+    } while (maxBufferSize > minSize && success == RTI_FALSE);
+
+    std::cout << "Maximum SHMEM allocable size: " << maxBufferSize << " Bytes." << std::endl;
```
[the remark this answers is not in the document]
salvacorts: Good point :+1:

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+
+    /** From user manual p780:
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+     * To optimize memory usage, specify a receive queue size less than that required to hold the maximum
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+     * at least 10 messages, then 640K bytes of memory would be needed if all messages were 64K bytes.
+     * Should this be desired, then receive_buffer_size should be set to 640K bytes.
+     *
+     * However, if the average message size is only 10K bytes, then you could set the receive_buffer_size to
+     * 100K bytes. This allows you to optimize the memory usage of the plugin for the average case and
+     * yet allow the plugin to handle the extreme case.
+     */
+    std::ostringstream ss;
+    const int perftest_overhead = 27;
+    const int rtps_overhead = 512;
```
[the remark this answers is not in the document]
salvacorts: Thanks. Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
         DDS_DomainParticipantQos& qos,
         ParameterManager *_PM)
 {
-    // Number of messages that can be buffered in the receive queue.
-    int received_message_count_max = 1024 * 1024 * 2
-            / (int) _PM->get<unsigned long long>("dataLen");
-    if (received_message_count_max < 1) {
-        received_message_count_max = 1;
-    }
+    DDS_Property_t *parentProp = DDSPropertyQosPolicyHelper::lookup_property(qos.property,
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+     * So for example, if the maximum message size is 64K bytes, and you configure the plugin to buffer
+     * at least 10 messages, then 640K bytes of memory would be needed if all messages were 64K bytes.
+     * Should this be desired, then receive_buffer_size should be set to 640K bytes.
+     *
+     * However, if the average message size is only 10K bytes, then you could set the receive_buffer_size to
+     * 100K bytes. This allows you to optimize the memory usage of the plugin for the average case and
+     * yet allow the plugin to handle the extreme case.
+     */
+    std::ostringstream ss;
+    const int perftest_overhead = 27;
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+    const int rtps_overhead = 512;
+    std::string flow_controller = _PM->get<std::string>("flowController");
+    // int flow_controller_token_size = atoi(
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+    unsigned long long int rtps_messages_per_sample = std::max(
+            1, (perftest_overhead + datalen) / fragment_size);
+
+    unsigned long long int received_message_count_max =
+            2 * (_PM->get<int>("sendQueueSize") + 1) * rtps_messages_per_sample;
 
-    std::ostringstream string_stream_object;
-    string_stream_object << received_message_count_max;
+    // min(maxBufferSize, received_message_count_max * rtps_message_size)
+    unsigned long long int receive_buffer_size = std::min(
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.h
```diff
@@ -16,6 +18,12 @@
   #include "PerftestSecurity.h"
 #endif
 
+/* Undefine Windows max and min micros.
```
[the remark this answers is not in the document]
salvacorts: Nice solution, thanks!

srcCpp/perftest_publisher.cxx
```diff
@@ -1691,19 +1830,19 @@ int perftest_cpp::Publisher()
...
+    * memory allocations in the critical path are done before the test begings,
+    * for both the Writer and the Reader that receives the samples.
+    * It will also serve to make sure that all the instances are registered
+    * in advance in the subscriber application.
+    *
+    * We query the MessagingImplementation class to get the suggested sample
+    * count that we should send. This number might be based on the reliability
+    * protocol implemented by the middleware behind. Then we choose between that
+    * number and the number of instances to be sent.
+    */
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/PerftestTransport.h
```diff
@@ -11,6 +11,12 @@
 #include <sstream>
 #include "ParameterManager.h"
 
+/* Undefine Windows max and min micros.
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/RTIDDSImpl.cxx
```diff
@@ -347,9 +381,14 @@ std::string RTIDDSImpl<T>::PrintConfiguration()
...
-    stringStream << "\n"
-                 << _transport.printTransportConfigurationSummary();
+    DDS_DomainParticipantQos *qos = new DDS_DomainParticipantQos;
+    if (this->_participant->get_qos(*qos) != DDS_RETCODE_OK) {
+        fprintf(stderr, "Could not get Participant QoS\n");
+    }
+
+    stringStream << "\n" << _transport.printTransportConfigurationSummary(qos);
 
+    if (qos != NULL) delete qos;
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/RTIDDSImpl.cxx
```diff
@@ -314,6 +324,30 @@ std::string RTIDDSImpl<T>::PrintConfiguration()
...
   #endif
 
+  #ifdef RTI_FLATDATA_AVAILABLE
+    // FlatData
+    stringStream << "\tFlatData: ";
+    if (_PM->get<bool>("flatdata")) {
+        stringStream << "Yes\n";
+    } else {
+        stringStream << "No\n";
+    }
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/RTIDDSImpl.cxx
```diff
@@ -722,177 +648,404 @@ class RTIPublisher : public IMessagingWriter
...
-        // Initialize data
-        if (!initialize_custom_type_dynamic_data(data)) {
-            // ~DDS_DynamicData() will be called and data will be free
-            throw std::runtime_error("initialize_custom_type_dynamic_data failed");
-        }
-      #endif
-        _PM = PM;
-        _writer = DDSDynamicDataWriter::narrow(writer);
-        DDS_Octet key_octets[KEY_SIZE];
+        this->_writer = T::DataWriter::narrow(writer);
```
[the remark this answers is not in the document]
salvacorts: This is accessing the parent class

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2078,9 +2349,7 @@ unsigned long RTIDDSImpl<T>::GetInitializationSampleCount()
      *
      * If we are not using reliability this should not matter.
      */
-    initializeSampleCount = (std::max)(
-            initializeSampleCount,
-            (unsigned long) _PM->get<int>("sendQueueSize"));
+    unsigned long initializeSampleCount = this->_sendQueueSize;
```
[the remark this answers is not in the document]
salvacorts: Because the actual send queue size can be modified for FlatData in large data.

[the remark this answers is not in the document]
salvacorts:
```C++
initial_samples = (std::max)(
                    1ul, (max_allocable_space - RTI_FLATDATA_MAX_SIZE) / RTI_FLATDATA_MAX_SIZE);

initial_samples = (std::min)(
                    initial_samples,
                    (unsigned long long) dw_qos.resource_limits.initial_samples);

dw_qos.resource_limits.initial_samples = initial_samples;
this->_sendQueueSize = initial_samples;
```

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2475,7 +2759,7 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
   #ifndef RTI_MICRO
     dw_qos.resource_limits.initial_instances = _PM->get<long>("instances") + 1;
 
-    if (_PM->get<int>("unbounded") != 0) {
+    if (_PM->get<int>("unbounded") != 0 && !_isFlatData) {
         char buf[10];
         sprintf(buf, "%d", _PM->get<int>("unbounded"));
```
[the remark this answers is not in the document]
salvacorts: Follow resolution on previous comment.

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2495,6 +2779,369 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
...
+
+          #ifdef RTI_DARWIN
+            /**
+             * In OSX, we might not be able to allocate all the send queue samples
+             * We only need this on the DW since it will allocate the samples
+             * on Zero Copy
+             */
+            if (_isZeroCopy) {
+                max_allocable_space = getShmemSHMMAX();
+                fprintf(stdout, "SHMMAX: %lu\n", max_allocable_space);
```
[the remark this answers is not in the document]
salvacorts: Yes, debugging.

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2388,6 +2670,7 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
         }
       #else
         dw_qos.resource_limits.max_samples = _PM->get<int>("sendQueueSize");
+        this->_sendQueueSize = dw_qos.resource_limits.max_samples;
```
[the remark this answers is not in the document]
salvacorts: Resolved on previous comment

srcCpp/RTIDDSImpl.cxx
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2495,6 +2779,369 @@ IMessagingWriter *RTIDDSImpl<T>::CreateWriter(const char *topic_name)
     }
   #endif
 
+  #ifdef RTI_FLATDATA_AVAILABLE
+    if (_isFlatData) {
+        char buf[10];
+        sprintf(buf, "%d", DDS_LENGTH_UNLIMITED);
+        DDSPropertyQosPolicyHelper::add_property(dw_qos.property,
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/perftest_publisher.cxx
```diff
@@ -154,15 +154,63 @@ int perftest_cpp::Run(int argc, char *argv[])
     } else {
         if (_PM.get<int>("unbounded") == 0) {
             if (_PM.get<bool>("keyed")) {
+              #ifdef RTI_FLATDATA_AVAILABLE
```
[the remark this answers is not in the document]
salvacorts: Wow, I like it lots. I was also thinking about a more elegant approach but this is pretty neat.

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
alfonso-soto: Or on the contractor `perftest_cpp::perftest_cpp()` where ever you prefer. Leave a comment about why we are initializing the seed.

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
alfonso-soto: What is the reason behind setting the `maxBufferSize` at 58MB?

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
alfonso-soto: I see, Ok thanks.

srcCpp/RTIDDSImpl.cxx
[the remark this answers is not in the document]
alfonso-soto: :+1:

srcCpp/RTIDDSImpl.cxx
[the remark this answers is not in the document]
alfonso-soto: :+1:

srcCppCommon/ParameterManager.cxx
alfonso-soto: Agree. :+1:

srcCpp/RTIDDSImpl.cxx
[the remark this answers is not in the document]
alfonso-soto: Nice

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
 {
-    // Number of messages that can be buffered in the receive queue.
-    int received_message_count_max = 1024 * 1024 * 2
-            / (int) _PM->get<unsigned long long>("dataLen");
-    if (received_message_count_max < 1) {
-        received_message_count_max = 1;
-    }
+    DDS_Property_t *parentProp = DDSPropertyQosPolicyHelper::lookup_property(qos.property,
+                "dds.transport.shmem.builtin.parent.message_size_max");
+    int parent_msg_size_max = atoi(parentProp->value);
```
alfonso-soto: Follow naming convention:
`parent_msg_size_max` -> `parentMsgSizeMax`

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
alfonso-soto: What's happens if the `"dds.transport.shmem.builtin.parent.message_size_max"` is higher than 58MB?

srcCpp/Infrastructure_pro.cxx
alfonso-soto: Apply also for the rest of the names.

srcCpp/Infrastructure_pro.cxx
```diff
@@ -501,21 +501,118 @@ bool configureShmemTransport(
...
+     * runtime errors.
+     *
+     * Perform an incremental search of allocable space in range (minSize, maxSize).
+     */
+    RTIOsapiSharedMemorySegmentHandle handle;
+    RTI_UINT64 pid = RTIOsapiProcess_getId();
+    RTIBool success = RTI_FALSE;
+    int retcode;
+    int key = rand();
+    int minSize = parent_msg_size_max;
```
alfonso-soto: It's more intuitive if you have the format for both of the variables on the range
`minBufferSize` and `maxBufferSize`

srcCpp/perftest_publisher.cxx
```diff
@@ -1306,32 +1388,89 @@ class LatencyListener : public IMessagingCB
         fflush(stdout);
 
       #ifndef RTI_MICRO
-        if (!unbounded) {
-            if (isKeyed) {
-                serializeTime = RTIDDSImpl<TestDataKeyed_t>::
-                        obtain_dds_serialize_time_cost(totalSampleSize);
-                deserializeTime = RTIDDSImpl<TestDataKeyed_t>::
-                        obtain_dds_deserialize_time_cost(totalSampleSize);
+        if (!isFlatData)  {
```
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
salvacorts: > Thanks for the explanation. What is the reason behind setting the maxBufferSize at 58MB?

58MB is used on DDS examples for FlatData and is enough to get maximum performance.

> What's happens if the "dds.transport.shmem.builtin.parent.message_size_max" is higher than 58MB?

`dds.transport.shmem.builtin.parent.message_size_max` Can it ever happen?

srcCpp/Infrastructure_pro.cxx
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
salvacorts: Done

srcCpp/Infrastructure_pro.cxx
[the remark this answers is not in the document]
alfonso-soto: I don't think so. :+1:

srcCpp/RTIDDSImpl.cxx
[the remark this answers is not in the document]
alfonso-soto: I just leave this here. https://stackoverflow.com/questions/4010281/accessing-protected-members-of-superclass-in-c-with-templates

Both solutions are right for me.

PR #205, opened 2019-07-23
Feature/fix rawtransport finalize

salvacorts:


srcCpp03/perftest_publisher.cxx
```diff
@@ -1019,6 +1018,11 @@ int perftest_cpp::RunSubscriber()
 
     perftest_cpp::MilliSleep(1000);
 
+    // TODO: Finalize receiverThread
```
javmorcas: I think you can remove the TODO

salvacorts: Done

salvacorts:
It is done

Issue #208, opened 2019-07-30
Issues in testing serialization/deserialization in 6.0.0


javmorcas:
**Description**

```
DDS_OctetSeq_set_member_element_count:ERROR: Failed to set sequence maximum
RTIXCdrInterpreter_fastDeserializeSample:TestDataLarge_t:bin_data deserialization error
Fail to deserialize sample on obtain_dds_deserialize_time_cost
DDS_OctetSeq_set_maximum:!assert buffer must not be loaned
DDS_OctetSeq_set_member_element_count:ERROR: Failed to set sequence maximum
RTIXCdrInterpreter_fastDeserializeSample:TestDataLarge_t:bin_data deserialization error
Fail to deserialize sample on obtain_dds_deserialize_time_cost
DDS_OctetSeq_set_maximum:!assert buffer must not be loaned
DDS_OctetSeq_set_member_element_count:ERROR: Failed to set sequence maximum
RTIXCdrInterpreter_fastDeserializeSample:TestDataLarge_t:bin_data deserialization error
Fail to deserialize sample on obtain_dds_deserialize_time_cost
DDS_OctetSeq_set_maximum:!assert buffer must not be loaned
DDS_OctetSeq_set_member_element_count:ERROR: Failed to set sequence maximum
RTIXCdrInterpreter_fastDeserializeSample:TestDataLarge_t:bin_data deserialization error
Fail to deserialize sample on obtain_dds_deserialize_time_cost
DDS_OctetSeq_set_maximum:!assert buffer must not be loaned
DDS_OctetSeq_set_member_element_count:ERROR: Failed to set sequence maximum
RTIXCdrInterpreter_fastDeserializeSample:TestDataLarge_t:bin_data deserialization error
Fail to deserialize sample on obtain_dds_deserialize_time_cost
```

**TODO**:
- [x] Support for C++ Traditional
- [x] Documentation
- [x] Full [Code Review]()
- [x] Final Merge into master

PR #209, opened 2019-08-05
Deserialize errors for large data

javmorcas:
This pull request is coming from the issue #208.

### DIFF
--- a/srcCpp/RTIDDSImpl.cxx
+++ b/srcCpp/RTIDDSImpl.cxx
@@ -2108,6 +2108,7 @@ double RTIDDSImpl<T>::obtain_dds_serialize_time_cost(
     double timeInit = 0;
     double timeFinish = 0;
     bool success = true;
+    unsigned int sequenceSize = sampleSize - perftest_cpp::OVERHEAD_BYTES;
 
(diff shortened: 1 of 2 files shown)


javmorcas:
@fercs77 @fercsrti77, could you check if this is correct? Is it ok to use `data.bin_data.ensure_length` instead of `data.bin_data.loan_contiguous`

srcCpp/RTIDDSImpl.cxx
```diff
@@ -2126,24 +2127,17 @@ double RTIDDSImpl<T>::obtain_dds_serialize_time_cost(
...
     /* --- Initialize data --- */
-    data.bin_data.maximum(0);
     data.entity_id = 0;
     data.seq_num = 0;
     data.timestamp_sec = 0;
     data.timestamp_usec = 0;
     data.latency_ping = 0;
-
-    /* Does not matter what it's inside */
-    data.bin_data.loan_contiguous(
```
javmorcas: File a bug about this.

javmorcas: It is not possible to deserialize a type containing an unbounded seq using a loaned buffer with enough size. This was not happening in 5.3.1

PR #219, opened 2019-08-28
Properly cast float to integer

salvacorts:


javmorcas:
@salvacorts Good catch!

PR #241, opened 2019-10-24
Fix ctf range bug

### LINKED ISSUE #240: Content filtering a range receives no samples by default
**Description**
When attempting to use -cft if you specify a range of IDs the subscriber will not receive any information without additional command line params
**How to Reproduce**
Using stock config files
pub side:
-pub -instances 10 -keyed
sub side:
-sub-keyed -cft 2:4

**Expected behavior**
sub should receive instances 2, 3, and 4
Single instance filtering like "-cft 4" will work
Or specifying -writeInstance 10 on the pub side will also force it to work

I would also recommend updating the documentation in 6.2.2.2
https://community.rti.com/static/documentation/perftest/3.0/examples.html#command-line-parameters
1: I was initially confused about the difference between -instances and -writeInstance, I recommend some clarification here that would help
2: I recommend mentioning you must use -keyed (this is probably obvious but I initially thought keyed instances would automatically be enabled if I specify -instances (maybe produce a warning if you don't use -keyed but you do use -cft or -instances to make it more obvious)

**TODO**:
- [x] Support for C++ Traditional
- [x] First CR [Code Review]()
- [x] Support for C++ Modern
- [x] Support for Java
- [x] Support for CS
- [x] Automatic install testing
- [x] Documentation
- [ ] Full [Code Review]()
- [ ] Final Merge into master

alfonso-soto:
This pull request is coming from the issue #240 . When the pull request is merged into "master" the issue will be fixed #240


alfonso-soto:
Maybe I have refactored too much.
This could be fixed with just a `i % KEY_SIZE`

srcCpp/RTIDDSImpl.cxx
```diff
@@ -3435,37 +3435,45 @@ DDSTopicDescription *RTIDDSImpl<T>::CreateCft(
...
+
+    /* Reserve memory for the range case, the bigger one. */
+    char cft_param[2 * KEY_SIZE][128];
+    const char* param_list[] = { cft_param[0], cft_param[1],
+            cft_param[2], cft_param[3],cft_param[4],
+            cft_param[5], cft_param[6], cft_param[7]
+    };
+
+    /* Only one element, no range */
+    if (cftRange.size() == 1) {
```
javmorcas: Kudos, this looks way cleaner!

javmorcas:
@alfonso-soto, I reviewed the code and added the documentation. If you agree, we will be ready to merge.

Issue #265, opened 2019-11-18
Rework idls to handle better custom types and Flat Data


javmorcas:
- Modify idl not to use sequences when using customTypes or customTypesFlatData
- Modify idl to set RTI_FLATDATA_MAX_SIZE directly there and have the option of overriding it via build script.

PR #268, opened 2019-11-19
#265 Add changes for not using bin_seq when doing custom Data

javmorcas:
This pull request is coming from the issue #265.


srcCpp/RTIDDSImpl.cxx
```diff
@@ -3795,9 +3799,12 @@ double RTIDDSImpl_FlatData<T>::obtain_dds_serialize_time_cost_override(
         builder.add_latency_ping(0);
 
         // Add payload
+        
+      #ifndef RTI_CUSTOM_TYPE_FLATDATA
```
salvacorts: Shouldn't we populate the custom type instead?

Otherwise having this metric when using custom types does not make much sense.

PR #274, opened 2019-11-28
Changes to algorithm that calculates SHMEM transport configuration pa…

fercsrti77:
I have created a pull request for the changes to the SHMEM configuration


srcCpp/Infrastructure_pro.cxx
```diff
@@ -539,8 +539,8 @@ bool configureShmemTransport(
     int retcode;
     int key = rand();
     int minBufferSize = parentMsgSizeMax;
-    int step = 1048576 + parentMsgSizeMax; // 1MB + parentMsgSizeMax
-    int maxBufferSize = 60817408; // 58MB
+    int step = 1048576; // 1MB
+    int maxBufferSize = (std::max)(60817408 /* 58MB */, parentMsgSizeMax);
```
javmorcas: I need to find out why we choose 58MB, this might look a bit random, and might need to be changed or explained. @fercsrti77, any feedback on this?

srcCpp/Infrastructure_pro.cxx
```diff
@@ -539,8 +539,8 @@ bool configureShmemTransport(
     int retcode;
```
javmorcas: Refered to line 511:

```
    DDS_Property_t *parentProp =
            DDSPropertyQosPolicyHelper::lookup_property(qos.property,
                    "dds.transport.shmem.builtin.parent.message_size_max");

    int parentMsgSizeMax = DEFAULT_MESSAGE_SIZE_MAX;
    if (parentProp != NULL && parentProp->value != NULL) {
        parentMsgSizeMax = atoi(parentProp->value);
    }
```

I believe, since you modified the place where this function is called, that the transport.messageSizeMax value is already populated. I believe we should use that one, instead of the message_size_max for just SHMEM, it will simplify code, plus, if you are setting SHMEM + UDPv4, the calculations will be the right ones.

srcCpp/Infrastructure_pro.cxx
```diff
@@ -822,6 +833,26 @@ bool PerftestConfigureTransport(
...
+    configureMessageSizeMaxTransport(transport, qos);
+
+    switch (transport.transportConfig.kind) {
+
+        case TRANSPORT_UDPv4:
+        case TRANSPORT_UDPv6:
+            break;
+
+        case TRANSPORT_SHMEM:
             if (!configureShmemTransport(transport, qos, _PM)) {
```
javmorcas: Should we maybe do this only if the 2 properties are not set via code? Like:

if (DDSPropertyQosPolicyHelper::lookup_property(qos.property,
                    "dds.transport.shmem.builtin.parent.received_message_count_max") == NULL
            && DDSPropertyQosPolicyHelper::lookup_property(qos.property,
                    "dds.transport.shmem.builtin.parent.receive_buffer_size") == NULL) {
    if (!configureShmemTransport(transport, qos, _PM)) {
                fprintf(stderr,
                        "%s Failed to configure SHMEM plugin\n",
                        classLoggingString.c_str());
        return false;
    }
}
break;

srcCpp/Infrastructure_pro.cxx
```diff
@@ -538,9 +570,9 @@ bool configureShmemTransport(
     RTIBool success = RTI_FALSE;
     int retcode;
     int key = rand();
-    int minBufferSize = parentMsgSizeMax;
-    int step = 1048576 + parentMsgSizeMax; // 1MB + parentMsgSizeMax
-    int maxBufferSize = 60817408; // 58MB
+    int minBufferSize = 1048576;
```
javmorcas: This looks wrong to me. This should be the minimum size of a message, not 1MB

javmorcas: @fercsrti77, I am changing this myself prior to the merge, lets review it later.

PR #275, opened 2019-11-29
Feature/267/not use max synchronous size new shmem adjustment

fercsrti77:
This pull request is coming from the issue #XX. When the pull request is merged into "master" the issue will be [resolved, closed, fixed] #XX

### DIFF
--- a/srcCpp/Infrastructure_pro.cxx
+++ b/srcCpp/Infrastructure_pro.cxx
@@ -585,10 +585,6 @@ bool configureShmemTransport(
     int step = 1048576; // 1MB
     int maxBufferSize = (std::max)(60817408 /* 58MB */, parentMsgSizeMax);
 
-    if (!messageSizeMaxSet) {
-
-    }
-
     do {
         // Reset handles to known state
         RTIOsapiMemory_zero(&handle,
--- a/srcCpp/Infrastructure_pro.cxx
+++ b/srcCpp/Infrastructure_pro.cxx
@@ -524,7 +524,7 @@ bool configureShmemTransport(
         datalen = scanList[scanList.size() - 1];
     }
 
-    int parentMsgSizeMax = DEFAULT_MESSAGE_SIZE_MAX;
+    int parentMsgSizeMax = transport.minimumMessageSizeMax;
     std::ostringstream ss;
     bool messageSizeMaxSet = false;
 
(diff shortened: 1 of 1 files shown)


srcCpp/Infrastructure_pro.cxx
```diff
@@ -580,10 +581,14 @@ bool configureShmemTransport(
     RTIBool success = RTI_FALSE;
     int retcode;
     int key = rand();
-    int minBufferSize = parentMsgSizeMax;
+    int minBufferSize = 1048576;
     int step = 1048576; // 1MB
     int maxBufferSize = (std::max)(60817408 /* 58MB */, parentMsgSizeMax);
 
+    if (!messageSizeMaxSet) {
```
javmorcas: What does this mean?

srcCpp/Infrastructure_pro.cxx
```diff
@@ -537,9 +537,12 @@ bool configureShmemTransport(
...
-    } else if (qos.transport_builtin.mask == DDS_TRANSPORTBUILTIN_SHMEM) {
-        if ((datalen + MESSAGE_OVERHEAD_BYTES) > DEFAULT_MESSAGE_SIZE_MAX) {
-            parentMsgSizeMax = datalen + MESSAGE_OVERHEAD_BYTES;
+    } else {
+        if (qos.transport_builtin.mask == DDS_TRANSPORTBUILTIN_SHMEM) {
+            if ((datalen + MESSAGE_OVERHEAD_BYTES) > parentMsgSizeMax) {
+                parentMsgSizeMax = datalen + MESSAGE_OVERHEAD_BYTES;
+                transport.minimumMessageSizeMax = parentMsgSizeMax;
+            }
         }
```
javmorcas: ok. I agree with this part so far. Just to clarify though:

If you are setting the property for SHMEM but it is SHMEM + something, then you use the value of the property. If you are NOT setting the property and it is SHMEM + something, then you use whatever is the minimum for the max_size across all transports.

That kind of makes sense, but it is also a weird scenario.

srcCpp/Infrastructure_pro.cxx
```diff
@@ -580,10 +581,14 @@ bool configureShmemTransport(
     RTIBool success = RTI_FALSE;
     int retcode;
     int key = rand();
-    int minBufferSize = parentMsgSizeMax;
+    int minBufferSize = 1048576;
     int step = 1048576; // 1MB
     int maxBufferSize = (std::max)(60817408 /* 58MB */, parentMsgSizeMax);
 
+    if (!messageSizeMaxSet) {
+
```
javmorcas: I am guessing this is in the wrong place, right? We are missing after the do-while a check with something like:

If size greater than 64k and I couldn't reserve memory for doing synchronous, default all the settings and try enabling asynchronous. Right?

srcCpp/Infrastructure_pro.cxx
fercsrti77: If you set the property "dds.transport.shmem.builtin.parent.message_size_max" for SHMEM we always use the property. 

If you don't set the property we only set it if the only transport is SHMEM

srcCpp/Infrastructure_pro.cxx
fercsrti77: In fact, that if had to be removed all together. Below we will adjust the transport size if messageSizeMaxSet is not true

srcCpp/Infrastructure_pro.cxx
javmorcas: That makes sense to me.

Issue #277, opened 2019-12-05
Building to use CustomType doesn't work anymore on windows


howard-rti:
**Description**
In trying out the example in 

https://github.com/rticommunity/rtiperftest/blob/master/srcDoc/examples.rst#using-custom-types
[m_custom.idl.txt](https://github.com/rticommunity/rtiperftest/files/3928615/m_custom.idl.txt)

It produces a build error:

C:\Users\howard\Desktop\rtiperftest>build.bat --platform x64Win64VS2015 --customType Test --customTypeFlatData Test_F
================================== PERFTEST: ===================================
[WARNING]: javac executable not found, perftest_java will not be built.
[WARNING]: PERL not found, C:\Users\howard\Desktop\rtiperftest\srcCpp\qos_string.h will not be updated.
[WARNING]: PERL not found, C:\Users\howard\Desktop\rtiperftest\srcCpp03\qos_string.h will not be updated.

[INFO]: Compilation flag for msbuild is: release
[INFO] FlatData is available
C:\Users\howard\Desktop\rtiperftest\srcIdl\customType\custom.idl
C:\Users\howard\Desktop\rtiperftest\srcIdl\customType\m_custom.idl
        2 file(s) copied.
INFO com.rti.ndds.nddsgen.Main Running rtiddsgen version 3.0.0, please wait ...
INFO com.rti.ndds.nddsgen.Main Done
INFO com.rti.ndds.nddsgen.Main Running rtiddsgen version 3.0.0, please wait ...
INFO com.rti.ndds.nddsgen.Main Done

[...]

PR #278, opened 2019-12-10
Hotfix custom types zero Copy on Windows

javmorcas:
This pull request is coming from the issue #277.


javmorcas:
This can be implemented in a different way more efficient

Issue #293, opened 2020-03-09
Repair link references in tutorial doc


pparrilla:
Compilation and download link in (performance_validation)[https://github.com/rticommunity/rtiperftest/blob/master/srcDoc/tutorials/performance_validation.rst] file need to change the link reference.

PR #294, opened 2020-03-09
Change incorrect links in documentation

pparrilla97:
This pull request is coming from the issue #293.


javmorcas:
@pparrilla97 Usually we would add to the Release notes a section explaining the changes, but in this case I believe we are fine with just changing the incorrect links. I'll go ahead and merge

PR #328, opened 2020-06-04
PerftestPrinter: Use a base class and inheritance to improve extensibility to other formats

javmorcas:

### DIFF
--- a/srcCpp/perftest_cpp.h
+++ b/srcCpp/perftest_cpp.h
@@ -52,7 +52,7 @@ class perftest_cpp
 
     // Private members
     ParameterManager _PM;
-    PerftestPrinter _printer;
+    PerftestPrinter *_printer;
     unsigned long long _SpinLoopCount;
     unsigned long _SleepNanosec;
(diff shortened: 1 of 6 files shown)


srcCpp/perftest_cpp.h
```diff
@@ -52,7 +52,7 @@ class perftest_cpp
 
     // Private members
     ParameterManager _PM;
-    PerftestPrinter _printer;
+    PerftestPrinter *_printer;
```
javmorcas, line 55: We need to use pointers in order to be able to choose later the right object

srcCpp/perftest_publisher.cxx
```diff
@@ -221,6 +219,16 @@ int perftest_cpp::Run(int argc, char *argv[])
         return -1;
     }
 
+    std::string outputFormat = _PM.get<std::string>("outputFormat");
+    if (outputFormat == "csv") {
+        _printer = new PerftestCSVPrinter();
```
javmorcas, line 224: This causes that we don't need the enum anymore

PR #355, opened 2020-10-19
PERF-194 -- Compilation issues for RTI Micro and QNX

javmorcas:
This pull request is coming from the RTI's Jira issue PERF-194.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -17,6 +17,7 @@ set "qos_file=%script_location%perftest_qos_profiles.xml"
 @REM # By default we will build pro, not micro
 set BUILD_MICRO=0
 set CMAKE_GENERATOR="NMake Makefiles"
+set ADDITIONAL_CMAKE_ARGS=""
 
 @REM # In case we build micro, which version.
 set BUILD_MICRO_24x_COMPATIBILITY=0
@@ -158,6 +159,9 @@ if NOT "%1"=="" (
 				SHIFT
 		) ELSE if "%1"=="--cmake-generator" (
 				SET "CMAKE_GENERATOR=%2"
+				SHIFT
+		) ELSE if "%1"=="--add-cmake-args" (
+				SET "ADDITIONAL_CMAKE_ARGS=%2"
 				SHIFT
 		) ELSE if "%1"=="--customType" (
 				SET USE_CUSTOM_TYPE=1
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -114,6 +115,9 @@ function usage()
     echo "                                 parameter is not present, cmake executable     "
     echo "                                 should be available from your \$PATH variable. "
     echo "                                 will clean all the generated code and binaries "
+    echo "    --add-cmake-args <s>         Additional defines and arguments that need to  "
```
Iceegan, line 118: I think this would be more clear if it said "Additional defines and arguments that *will be* passed to the cmake executable..." Since on many platforms you will not need these arguments

build.bat
```diff
@@ -984,6 +988,9 @@ GOTO:EOF
 	echo.    --cmake  path                Path to the CMAKE executable. If this
 	echo.                                 parameter is not present, Cmake variable
 	echo.                                 should be available from your $PATH variable.
+	echo.    --add-cmake-args <s>         Additional defines and arguments that need to
```
Iceegan, line 991: Same as build.sh: I think this would be more clear if it said "Additional defines and arguments that will be passed to the cmake executable..." Since on many platforms you will not need these arguments. I could see "need to" causing confusion on whether this is a required argument.

srcDoc/release_notes.rst
```diff
@@ -193,6 +193,15 @@ In previous releases, the `qos_string.h` file containing the strings from the
 `srcCpp03` folders. This process has been simplified, generating only once and
 into `srcCppCommon`.
 
+Add command-line argument to the `cmake` executable (PERF-194)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Starting in this release *RTI Perftest* enables the customer to add specific
+command-line arguments to the `cmake` executable (which is used when compiling
+against *RTI Connext DDS Micro*. This feature allows the control of the compiler
```
Iceegan, line 201: A small typo: there is a missing closing parenthesis after "when compiling against *RTI Connext DDS Micro*"

Iceegan (commented):
I think these changes are good. I only had a few suggestions about wording in the documentation. 

An additional change that may be worth adding (but would take more effort) would be to create and distribute a set of toolchain files for common platforms that can set the correct  CMake flags. This is what rtime-make does in Micro (see the toolchains in `<RTIMEHOME>/resource/cmake/architectures`). The build script could select the toolchain based on the --platform argument. We might not be able to write toolchain files for every platform, but there is an opportunity for the community to contribute ones for any platforms they have successfully compiled on.

javmorcas:
@Iceegan thanks for your feedback. I will add your suggestions to the code. I will also discuss with you further your suggestion about the architectures, maybe we can even read them from the path you gave me and use them to add the info to the cmake line.

build.sh
javmorcas: Done

build.bat
javmorcas: Done

srcDoc/release_notes.rst
javmorcas: Done

